PR #1, opened 2021-03-17
Clean up protos

azdagron:
- Fixed up go module paths to point to this repository instead of SPIRE
- Changed go package names for easier importing via `goimports` (i.e.`.../bundle/v1` imports as `bundlev1` instead of `bundle`)
- Renames spire/types ==> spire/api/types to cleanly namespace types related to the API
- Removed some bad trailing whitespace and mixed tabs/spaces


amartinezfayo (approved):
Thanks for this, @azdagron!
Changes look good to me. I see the renaming of `spire/types` to `spire/api/types` very beneficial to avoid confusions on the scope to use it.

PR #5, opened 2021-05-10
Federation Management API: define a Protobuf

amartinezfayo:
Added Protobuf definitions for API-based federation management.

Fixes spiffe/spire#2250.


proto/spire/api/server/federationmanagement/v1/federationmanagement.proto
```diff
@@ -0,0 +1,154 @@
+syntax = "proto3";
+package spire.api.server.federationmanagement.v1;
```
evan2645, line 2: I'm wondering about the API name ...

If we are going to stick to the "federation" term, it's hard for me to come up with something better than this ... but maybe there are other ways to look at it? For example, maybe trust domain API makes sense? Or bundle endpoint API?

I think part of our difficulty here is that with the other APIs, we are resource-centric ... but, for federation, we never really came up with a good name for this resource that is being managed. In the config, we used `federate_with` ... `federation relationship` could work, but I wonder what else we could call them.

proto/spire/api/server/federationmanagement/v1/federationmanagement.proto
```diff
@@ -0,0 +1,154 @@
...
+    //
+    // The caller must be local or present an admin X509-SVID.
+    rpc BatchDeleteFederationRelationship(BatchDeleteFederationRelationshipRequest) returns (BatchDeleteFederationRelationshipResponse);
+
+    // Refreshes the bundle from the specified federated trust domain.
+    // If there is not a federation relationship configured with the
+    // specified trust domain, NOT_FOUND is returned.
+    //
+    // The caller must be local or present an admin X509-SVID.
+    rpc RefreshBundle(RefreshBundleRequest) returns (google.protobuf.Empty);
```
evan2645, line 45: This one feels a little inconsistent with the rest. Maybe `ReloadFederationRelationship` could work?

evan2645: How do we envision this working? It will trigger a refresh/reload on the server handling this RPC? It will be blocking or non-blocking?

proto/spire/api/types/federationrelationship.proto
```diff
@@ -0,0 +1,39 @@
...
+message FederationRelationship {
+    // The trust domain name (e.g., "example.org") to federate with. 
+    string trust_domain = 1;
+
+    // URL of the SPIFFE bundle endpoint that provides the trust
+    // bundle to federate with. Must use the HTTPS protocol.
+    string bundle_endpoint_url = 2;
+
+    // Required. The endpoint profile type.
+    oneof bundle_endpoint_profile {
```
evan2645, line 14: 💯

proto/spire/api/types/federationrelationship.proto
```diff
@@ -0,0 +1,39 @@
...
+        // Use Web PKI endpoint profile. 
+        WebPKI https_web = 3;
+
+        // Use SPIFFE Authentication endpoint profile.
+        SPIFFEAuthentication https_spiffe = 4;
+    }
+}
+
+message SPIFFEAuthentication {
```
evan2645, line 23: Should we also have a bundle included in this message?

If I am creating a new relationship with SPIFFE auth ... I need to set the bundle first. Do we need to call bundle api, set the bundle ... _then_ call this api to set up the relationship?

I think this starts to tickle some of my earlier comments about understanding this resource a little better. It is closely related with the Bundles resource ... I do think there are use cases for bundles without federation, for example if you roll your own bundle distribution mechanism. But the relationship makes these API boundaries a little less clear. Curious to hear your thoughts on this

evan2645 (commented):
Thank you so much for this @amartinezfayo! I think it's looking great - I left a few naming comments/questions

proto/spire/api/server/federationmanagement/v1/federationmanagement.proto
amartinezfayo: >I think part of our difficulty here is that with the other APIs, we are resource-centric

Yes, that's a really good point. I've been thinking about this and Trust Domain API seems to be the most appropriate name IMO. I prefer Trust Domain API instead of Bundle Endpoint API because 1) We are really managing the relationships between trust domains, not the bundle endpoints themselves. 2) It's out of the scope of the API to manage the bundle endpoint server. The API deals with bundle endpoints as a way to manage relationships between trust domains.

proto/spire/api/server/federationmanagement/v1/federationmanagement.proto
amartinezfayo: >This one feels a little inconsistent with the rest. Maybe ReloadFederationRelationship could work?

The reason behind the name is that the primary goal is really to refresh the bundle from a specified trust domain. We currently log a message at Info level saying `Bundle refreshed` when a bundle has been successfully refreshed and seems natural to call it `RefreshBundle`.
I see your point about consistency with other RPCs, but my fear is that `ReloadFederationRelationship` may induce to think that there is an attribute of the federation relationship itself that you want to refresh, e.g. the bundle endpoint profile used or the URL and may not be obvious that the bundle will be refreshed.
Having said that, I'm not strongly opposed to change the name to `ReloadFederationRelationship` since it also has a logic.

amartinezfayo: >How do we envision this working? It will trigger a refresh/reload on the server handling this RPC? It will be blocking or non-blocking?

The use case that I saw for this RPC is: TD1 has a federation relationship with TD2. An operator or automated process in TD1 knows that the bundle of TD2 has changed for whatever reason and the next bundle refresh is scheduled to happen in days, which is not acceptable. The RPC is called in TD1 to refresh the bundle from TD2 which would effectively do the same as what would happen in the next scheduled refresh, i.e. set the updated bundle of TD2 in TD1. I envisioned this as a blocking call as an easy way to know that the operation completed, but I'm open to suggestions.

proto/spire/api/types/federationrelationship.proto
amartinezfayo: This is a good point. I was hesitant to include the bundle here but I was influenced by the current implementation where you are forced to set the federated bundles separately from the configuration of the federation relationships.
I think that having the possibility to include the bundle here will make things easier and avoid extra calls to set federated bundles separately, mostly in the "bootstrapping of federation" process.
I would make this optional so if you don't provide it, it looks for a bundle previously stored.

amartinezfayo: Specifying the bundle here makes a lot of sense in the case of chaining trust with SPIFFE Authentication, because you need the bundle of the trust domain where the bundle endpoint server resides and you may not need/have a federation relationship with that trust domain.

proto/spire/api/types/federationrelationship.proto
```diff
@@ -0,0 +1,46 @@
...
+}
+
+message HTTPSSPIFFEProfile {
+    // Specifies the expected SPIFFE ID of the SPIFFE bundle endpoint server.
+    string endpoint_spiffe_id = 1;
+
+    // The bundle of the trust domain that will be used when performing requests
+    // to the SPIFFE bundle endpoint server. It will be persisted for future requests.
+    // If there is an existing stored bundle, this bundle will be used instead and the
+    // existing bundle will be updated. If not specified, a previously
```
evan2645, line 32: i think this partial suggestion is more accurate? Or will we use the specified bundle only on the first invocation and not write it to the bundle table prior to that?

```suggestion
    // If there is an existing stored bundle, this bundle will overwrite it. If
    // not specified, a previously
```

evan2645 (approved):
This is looking good to me! Would like to get review from @azdagron or @rturner3 too if possible

azdagron (dismissed):
LGTM after addressing evan's suggestion!

proto/spire/api/types/federationrelationship.proto
amartinezfayo: Thanks for the suggestion, I meant to overwrite it :)

evan2645 (approved):
🎸 ♾️

azdagron (approved):
:shipit:

PR #6, opened 2021-05-26
Add match any and match superset

MarcosDY:
Add new Match when filtering BySelector and ByFederatesWith


proto/spire/api/types/selector.proto
```diff
@@ -20,6 +20,14 @@ message SelectorMatch {
         // Indicates that the selectors in this match are a subset of the
         // candidate selectors.
         MATCH_SUBSET = 1;
+
+        // Indicates that the selectors in this match are a superset of the
+        // candidate selectors.
```
evan2645, line 25: Hmm I think these comments imply the inverse of the comments in the FederatesWithMatch message. There, the term `candidate` is used for entries that will be returned by the query. Here, we are saying that MATCH_SUPERSET will return candidates that are a _subset_ of the provided selectors, and we say the opposite in FederatesWithMatch. FederatesWithMatch is the behavior we want.

MarcosDY: I apply some changes and add examples to all matches I think it can help us to avoid wording issues

evan2645 (approved):
Thanks @MarcosDY, this looks great ... the explanations make things very clear.

I'm sorry for the prolonged delay on this. The 1.0 release was not very fun, and this change fell victim to it

PR #8, opened 2021-07-22
proto/agent: add Delegated Identity API

rscampos:
Add proto definitions for new Delegated Identity API. This API includes two
rpcs: SubscribeToX509SVIDs and SubscribeToX509Bundles.

The SubscribeToX509SVIDs rpc allows a privileged client to get X509-SVIDs
for a given workload. This method uses a directional gRPC stream.

The request is used to subscribe a workload, and the response (stream
from server to client) is used to send SVID updates (for the subscribed
workload). The client subscription is based on the workload's selectors.
X509-SVIDs for registration entries matching *all* the passed selectors
are returned.

The SubscribeToX509Bundles rpc streams get local and all federated
bundles.

Fixes https://github.com/spiffe/spire-api-sdk/issues/4
Related https://github.com/spiffe/spire-api-sdk/pull/7

Co-authored-by: Raphael Campos <raphael@accuknox.com>
Co-authored-by: Thiago Navarro <navarro@accuknox.com>
Signed-off-by: Mauricio Vásquez <mauricio@accuknox.com>
Signed-off-by: Mauricio Vásquez <mauricio@kinvolk.io>
Signed-off-by: Raphael Campos <raphael@accuknox.com>
Signed-off-by: Thiago Navarro <navarro@accuknox.com>


proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
```diff
@@ -0,0 +1,59 @@
...
+// X.509 SPIFFE Verifiable Identity Document with the private key.
+message X509SVIDWithKey {
+    // The agent X509-SVID.
+    spire.api.types.X509SVID x509_svid = 1;
+
+    // Key encoding (DER PKCS#8).
+    bytes x509_svid_key = 2;
+
+    // Names of the trust domains that this workload should federates with.
+    repeated string federates_with = 3;
```
evan2645, line 36: In the workload api, we return this at the top level (i.e. as part of `SubscribeToX509SVIDsResponse`). The thinking was that we didn't want to slide into authorization land by telling workloads which trust domains a particular SVID should be used in conjunction with.

It's hard for me to tell whether that's the right thing to do here or not. There is also the question of consistency with the workload api, but we've already departed from that a bit.

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
```diff
@@ -0,0 +1,59 @@
...
+}
+
+message SubscribeToX509SVIDsResponse {
+    repeated X509SVIDWithKey x509_svids = 1;
+}
+
+message SubscribeToX509BundleRequest {}
+
+// SubscribeToX509BundleResponse represents the bundle for a given trust domain. An
+// empty bundle indicates that it was removed. Use DER-encoded in X.509 certs.
```
evan2645, line 53: > An empty bundle indicates that it was removed.

An alternative could be to have one message that sends all the bundles. When any of them change (or are removed), the entire set is retransmitted. This is closer to the behavior of the workload api (and the behavior of `SubscribeToX509SVIDs` I assume?), but also means the message size will be much larger. I'm not sure how easy it would be to start running into limitations there

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
```diff
@@ -0,0 +1,59 @@
...
+// workloads on the host. This service is intended for use cases where a process
+// (different than the workload one) should access the workload's SVID to
+// perform actions on behalf of the workload. One example of is using a single
+// node instance of Envoy that upgrades TCP connections for different processes
+// running in such a node.
+//
+// The caller must be local and its identity must be listed in the allowed
+// clients on the spire-agent configuration.
+service DelegatedIdentity {
+    // Subscribe to get X.509-SVIDs for identities that match the given selectors.
```
evan2645, line 18: Since we are delegating workload SVIDs here, I think it is natural to frame the comment as obtaining a workload's SVIDs rather than obtaining an identity's SVIDs

```suggestion
    // Subscribe to get X.509-SVIDs for workloads that match the given selectors.
```

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
```diff
@@ -0,0 +1,59 @@
...
+    // The lifetime of the subscription aligns to the lifetime of the stream.
+    rpc SubscribeToX509SVIDs(SubscribeToX509SVIDsRequest) returns (stream SubscribeToX509SVIDsResponse);
+
+    // Subscribe to get local and federated bundle.
```
evan2645, line 22:
```suggestion
    // Subscribe to get local and all federated bundles.
```

evan2645 (commented):
Thanks for cleaning this up @rscampos

I think it is looking awesome. No big structural stuff, I left mostly comments about comments 😆 

Behavior of bundle subscription, as well as where we put the federated bundle references, are the two "biggest" things. Curious to hear everyone's thoughts there ... hope to see this get merged soon

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
rscampos: Hey @evan2645, I took a step back to follow the discuss in old https://github.com/spiffe/spire-api-sdk/pull/7. If I misunderstanding something, pls let me know. 

Are you saying that, today a workload can receive all trust domains (in `federates_with`) which is able to federate with, and this is not a good way to slide into authorization land?
Let's say that an alternative way would be a workload asking if its could federate with a specific trust domain? And, in spire, a check, answering true/false for the trust domain informed?

I don't know if I catch the ideia. Do you think that a critical point to an initial version?

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
rscampos: @evan2645 , we discuss it and we think would be better to send all the bundles in one message too. About sending an empty message is it ok?
About the message size, would be it critical for now? what do you think?

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
evan2645: It is not a critical point. Since this API is not part of SPIFFE spec we can deprecate and change as needed.

Workload-to-identity is 1:N. The question here is whether federated_bundle should be in the workload or the identity.

I'm fine with this as-is, I just wanted to point it out in case someone had an opinion on it :)

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
evan2645: Re: message size ... the experience here will be, the integration will break entirely for any node that is running one or more federated workloads where the sum of the federated bundles across all workloads on the node is >~4MB. It can be that the addition of a new federated bundle causes us to cross this threshold (and fail suddenly). How likely this is, I'm not sure ...

We already have this problem on workload api, but it is on a per-workload basis. Here, we are talking about per-node. We should probably figure out message overhead and how many bytes per cert to better understand how realistic this worry is.

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
azdagron: I think I'd slightly prefer it top-level in the response instead of duplicated for each identity, for both consistency sake with the Workload API and for reasons even has mentioned (sliding into authorization land).

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
azdagron: I think I'd be fine with starting with a "send the world" approach to this at first. We can always add a different RPC later if we determine this is too problematic. I think the "send the world" approach is going to be simpler to implement on both sides, and is probably an ok place to start. Certainly we can imagine a world where the number of federation relationships is high enough to trip the message limit, but in most instances it feels contrived.

evan2645: > Certainly we can imagine a world where the number of federation relationships is high enough to trip the message limit, but in most instances it feels contrived.

How many certs do we think here to bust the limit? 100? 1,000? 10,000? The first doesn't seem impossible, the second is unlikely IMO. I think the important thing here is the failure mode and how we handle it. We probably don't want an outage being caused across the fleet because an operator added a new federation relationship.

azdagron: So, a cert over an EC256 public key with a SHA256 hash with minimal extensions is ~300 bytes. A cert over an RSA 2048-bit key of similar complexity is ~700 bytes (somewhat rough ballpark figures). We're probably looking on the order of thousands of certificates before tripping up.

evan2645: 👍 thanks @azdagron, sounds like we'll be ok for now.

if we're alright on message size, my vote is for one message with all the bundles.

azdagron: SGTM!

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
```diff
@@ -0,0 +1,59 @@
...
+message SubscribeToX509BundleRequest {}
+
+// SubscribeToX509BundleResponse represents the bundle for a given trust domain. An
+// empty bundle indicates that it was removed. Use DER-encoded in X.509 certs.
+message SubscribeToX509BundleResponse {
+    string trust_domain_name = 1;
+
+    // DER-encoded X.509 certs
+    repeated bytes ca_certificates = 2;
+}
```
evan2645, line 59:
```suggestion
message SubscribeToX509BundlesResponse {
    // A map keyed by trust domain name, with ASN.1 DER-encoded
    // X.509 CA certificates as the values
    map<string, bytes> ca_certificates = 1;
}
```

evan2645 (commented):
I've left some suggestions here to bring the PR up to the single-message-for-all-bundles shape that we've recently discussed. Please let me know what you think. If you agree with this change, I think it's the last outstanding thing! Thanks so much for hanging in there, I trust that we have spent our time here well.

azdagron:
Those suggestions look good to me.

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
rscampos: @azdagron In the case we put `federates_with` at top-level in the response (as the message `X509SVIDResponse`), we need to aggregate all the foreign trust domains (based on each registration entry) before sending them to workload, right?

azdagron: Yep.

rscampos: @azdagron @evan2645 maybe put the `federates_with` at top-level in the response make more sense once the Workload API do it in this way. At first I didn't get the ideia, but now I understand looking at `X509SVIDResponse`. By now I can't see a problem doing in this way :)
Tks again for all the discuss guys :D

rscampos:
@azdagron @evan2645
I sent a new commit with all new modifications. I also put the `federates_with` in top-level response. Please let me know if everything is correct.

Thank you a lot folks :D

evan2645 (approved):
Thanks so much everyone! I think we're ready to go here

PR #9, opened 2021-08-26
Add store_svid to entry type

MarcosDY:
Add  store_svid to type/entry, that will be used to determinate entries that are "storable"


azdagron (approved):
\o/

PR #10, opened 2021-09-02
Update trust domain proto messages to support static federation relationships

marcosy:
This PR updates the trustdomain proto messages to support static federation relationships according to the discussion in: [#2260](https://github.com/spiffe/spire/issues/2260)

Signed-off-by: Marcos Yedro <marcosyedro@gmail.com>

### DIFF
--- a/proto/spire/api/server/trustdomain/v1/trustdomain.proto
+++ b/proto/spire/api/server/trustdomain/v1/trustdomain.proto
@@ -56,6 +56,10 @@ message ListFederationRelationshipsRequest {
 
     // The next_page_token value returned from a previous request, if any.
     string page_token = 3;
+
+    // If true, static federation relationships defined in the SPIRE server
+    // configuration file are included in the response.
+    bool include_static = 4;
 }
 
 message ListFederationRelationshipsResponse {
@@ -73,9 +77,13 @@ message GetFederationRelationshipRequest {
     // (e.g., "example.org").
     string trust_domain = 1;
 
+    // If true, it returns the static federation relationship defined in the
+    // SPIRE server configuration file.
+    bool include_static = 2;
+
     // An output mask indicating which federation relationship fields
     // are set in the response.
-    spire.api.types.FederationRelationshipMask output_mask = 2;
+    spire.api.types.FederationRelationshipMask output_mask = 3;
 }
 
 message BatchCreateFederationRelationshipRequest {
--- a/proto/spire/api/types/federationrelationship.proto
+++ b/proto/spire/api/types/federationrelationship.proto
@@ -12,13 +12,18 @@ message FederationRelationship {
     // bundle to federate with. Must use the HTTPS protocol.
     string bundle_endpoint_url = 2;
 
+    // If true, it indicates that the federation relationship is static. Static
+    // federation relationships are those defined in the SPIRE server
+    // configuration file.
+    bool static = 3;
+
     // Required. The endpoint profile type.
     oneof bundle_endpoint_profile {
         // Use Web PKI endpoint profile. 
-        HTTPSWebProfile https_web = 3;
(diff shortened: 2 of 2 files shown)


proto/spire/api/server/trustdomain/v1/trustdomain.proto
```diff
@@ -73,9 +77,13 @@ message GetFederationRelationshipRequest {
...
     string trust_domain = 1;
 
+    // If true, it returns the static federation relationship defined in the
+    // SPIRE server configuration file.
+    bool include_static = 2;
+
     // An output mask indicating which federation relationship fields
     // are set in the response.
-    spire.api.types.FederationRelationshipMask output_mask = 2;
+    spire.api.types.FederationRelationshipMask output_mask = 3;
```
azdagron, line 86: Note for other reviewers. This API has not been published/implemented yet so there are no existing clients to break by renumbering the fields. I was just about to comment with a "NO, DON'T!" when I realized it myself :)

proto/spire/api/server/trustdomain/v1/trustdomain.proto
```diff
@@ -73,9 +77,13 @@ message GetFederationRelationshipRequest {
     // (e.g., "example.org").
     string trust_domain = 1;
 
+    // If true, it returns the static federation relationship defined in the
+    // SPIRE server configuration file.
+    bool include_static = 2;
+
```
azdagron, line 83: Hmm, I wonder about this behavior. If you call `Get` with `include_static=true` and no static entry exists, do you get back NOT_FOUND or the dynamic entry (assuming there is a dynamic entry)?

Sorry to rehash this all again, but I wonder.... should we just simply have separate RPCs for the static relationships? e.g., `GetStaticFederationRelationship` and `ListStaticFederationRelationships` (there would be no create/update/delete counterparts)

azdagron (commented):
Missing the codegen here... `make` needs to be run.

proto/spire/api/server/trustdomain/v1/trustdomain.proto
amartinezfayo: I thought about the NOT_FOUND case when considering the introduction of this field. I think that there are some pros / cons with each approach.
One thing that I don't love about having separate RPCs is that in that case you have to make multiple calls if you want to list the full set of relationships and join the results.
With a single RPC and a field you can get all the relationships in a single call but as you pointed the semantics is less clear in the NOT_FOUND case.

I don't really have a strong opinion. Just wanted to point one disadvantage of going with multiple RPCs. It also may not be super obvious that `ListFederationRelationships` will return dynamic entries only, by looking at its name. So we may need to rename the RPCs to reflect that they will deal with dynamic entries only.

marcosy: > If you call `Get` with `include_static=true` and no static entry exists, do you get back NOT_FOUND or the dynamic entry (assuming there is a dynamic entry)?

I would expect to get NOT_FOUND in that case.

I am ok with both approaches. If the concern is only about clarity, another option could be to use a `data_source` attribute for Get and List that could be set to `static`, `dynamic` and `static_and_dynamic` using an enum.

GetFederationRelationships would fail if invoked with `static_and_dynamic` (or one can take precedence over the other)

azdagron:
We talked about this IRL during the contributor sync meeting. @evan2645 suggested we punt on if/how static relationships are reflected in the API for the first release. The group agreed on that course of action.

marcosy:
Closing this PR according to what we discussed in contributor sync meeting. The current API design seems to cover well the dynamic federation relationships use case and also leaves enough room for introducing the static federation relationship management in the future in a back-compatible way. 

I don’t think any other change is needed at this point.

PR #14, opened 2021-09-24
Update generated code after dep bump

azdagron:
We bumped the deps in #13 but neglected to generate the code.

Also including a github action to check that the generated code is up to date.


evan2645 (approved):
Gotta love how easy actions are

PR #16, opened 2021-09-24
Update name/description of federation relationship bundle field

azdagron:
This brings the comments in line with the SPIRE server behavior and updates the field name to be more reflective of the purpose.


proto/spire/api/types/federationrelationship.proto
```diff
@@ -26,11 +26,13 @@ message HTTPSSPIFFEProfile {
     // Specifies the expected SPIFFE ID of the SPIFFE bundle endpoint server.
     string endpoint_spiffe_id = 1;
 
-    // The bundle of the trust domain that will be used when performing requests
-    // to the SPIFFE bundle endpoint server. It will be persisted for future requests.
-    // If there is an existing stored bundle, this bundle will overwrite it. If
-    // not specified, a previously stored bundle must exist.
-    spire.api.types.Bundle bundle = 2;
+    // The bundle used to bootstrap the connection to the SPIFFE bundle
+    // endpoint server. This field is required for self-serving endpoints (i.e.
```
evan2645, line 30: Is it required if a bundle is already set for the trust domain? E.g. Delete then re-create the relationship, or create the relationship then set later

azdagron: Currently, it is. Not sure it should be.

azdagron: The more I think about it, the less inclined I am to require the bundle in any situation. I'm ----><---- close to thinking maybe we just get rid of this field altogether but there is a convenience factor that maybe I'm not evaluating clearly.

azdagron: I've almost come full circle and want to get rid of the notion of self-serving v.s. non self-serving in our experience. I'm not sure what the spec would have to say about it, or if it is dangerous somehow. I'm leaning on two directions:

1. get rid of that field altogether. we can return a field in the response letting the caller know if the endpoint bundle is known or not when the relationship is made, or...
2. that field is named endpoint_bundle, (or endpoint_bootstrap_bundle), is optional, and if set, is used to create/replace the bundle for the endpoint trust domain. We would always return the endpoint bundle on a get/list if the bundle is known. (edited) 

We can always make the replace functionality opt-in or opt-out if we determine the correct default behavior but think it may not be universally wanted.

azdagron: We've discussed this in the SPIFFE slack. The updated change reflects the outcome.

proto/spire/api/types/federationrelationship.proto
```diff
@@ -20,17 +20,25 @@ message FederationRelationship {
         // Use SPIFFE Authentication endpoint profile.
         HTTPSSPIFFEProfile https_spiffe = 4;
     }
+
+    // Optional. The bundle for the trust domain. This field can be used to
+    // create/replace the bundle when the relationship is created or updated.
```
amartinezfayo, line 25: Since the field will be in the `FederationRelationship` message and will not be specific for the `https_spiffe` profile, I’m wondering if it’s clear that for create/update the field is only useful when using the `https_spiffe` profile. If I’m creating a relationship with the https_web profile I may think that I need to fill it.

So we may say something like
_This field can be used to create/replace the bundle when the relationship is created or updated using the https_spiffe profile_

What do you think?

azdagron: Due to it's generic nature, I'd prefer not to have special case behavior for this field based on the profile of choice. In my opinion, if supplied, no matter the profile, it should be create/update the bundle for the trust domain.

evan2645: > Since the field will be in the FederationRelationship message and will not be specific for the https_spiffe profile, I’m wondering if it’s clear that for create/update the field is only useful when using the https_spiffe profile.

...

> In my opinion, if supplied, no matter the profile, it should be create/update the bundle for the trust domain.
> In other words, this field is simply a convenience for setting the trust domain bundle for the relationship at the same time the relationship is created, instead of needing to use two calls.

I think there is some use (call it a small convenience) even for other profiles as it should allow your workloads to begin federating immediately vs waiting for the first bundle endpoint conversation to take place (which may even fail).

I think we also want this bundle to be reflected in API responses regardless of the profile details?

evan2645, line 25: nits
/ -> or
more bundle clarity

```suggestion
    // create or replace the referenced trust domain's bundle when the relationship is created or updated.
```

amartinezfayo (dismissed):

proto/spire/api/types/federationrelationship.proto
```diff
@@ -20,17 +20,25 @@ message FederationRelationship {
...
+    // When the relationship is retrieved, it will be set to the latest trust
```
evan2645, line 26: nit: which trust domain? ours or theirs? i know it sounds silly in the context, but the `bundle` term can never be harmed by over-clarification

```suggestion
    // When the relationship is retrieved, it will be set to the referenced trust domain's latest bundle
```

proto/spire/api/types/federationrelationship.proto
```diff
@@ -20,17 +20,25 @@ message FederationRelationship {
...
+    // domain bundle (if available). Note, an endpoint with an https_profile
```
evan2645, line 27:
```suggestion
    // domain bundle (if available). Note, an endpoint with an https_spiffe profile
```

proto/spire/api/types/federationrelationship.proto
```diff
@@ -20,17 +20,25 @@ message FederationRelationship {
...
+    // that is self-served (i.e. bundle endpoint server presents an SVID from
+    // the same trust domain as the bundle returned from the endpoint) requires
+    // the bundle from that trust domain to authenticate the connection. This
+    // parameter is a convenient way to provide that bundle instead of making a
+    // seperate call to the Bundle APIs. If the bundle for the endpoint does
+    // not exist, for both self-serving and non self-serving endpoints, the
+    // relationship will still be created.
```
evan2645, line 34: alternative: Here is a partially-suitable suggestion that I think may help to de-complicate some of the mental model around this

```suggestion
    // domain bundle (if available). Please note that the `https_spiffe` profile requires
    // an existing trust domain bundle in order to function correctly. The required bundle
    // must match the trust domain specified in the bundle endpoint SPIFFE ID. If the bundle
    // endpoint SPIFFE ID resides in the same trust domain that you're trying to federate with,
    // you may optionally specify it here. If the bundle endpoint SPIFFE ID _does not_ reside in
    // the same trust domain that you're federating with, please ensure that ..
```

evan2645 (dismissed):
I think we can improve the comment. I approved anyway because we have room to fix up the comment after the field name is updated ... but we should still fix up the comment before release.

amartinezfayo (approved):
LGTM!

PR #20, opened 2022-02-21
proto/agent: add JWT SVID privileged API

loveyana:
Add JWT SVID privileged API proto definitions in Delegated Identity API. This API includes two RPCs: `FetchJWTSVIDs` and `SubscribeToJWTBundles`.

The FetchJWTSVIDs rpc allows a privileged client to get JWT-SVIDs for a given workload. 

The request is used to fetch JWT-SVIDs for a workload. As in the workload api, audience list is required and also need the selectors describing the workload to fetch.

And the response is used to fetch a list of JWT-SVIDs (for the fetched workload).  JWT-SVIDs for registration entries matching all the passed selectors are returned.

The `SubscribeToJWTBundles` rpc streams get local and all federated bundles. It should be noted that the bundle data it returns is marshaled to JWKS.

Related: https://github.com/spiffe/spire-api-sdk/issues/4, https://github.com/spiffe/spire-api-sdk/pull/7 ,https://github.com/spiffe/spire-api-sdk/pull/8


loveyana:
@amartinezfayo I'm glad you noticed this, do you have any thoughts about it?

proto/spire/api/agent/delegatedidentity/v1/delegatedidentity.proto
```diff
@@ -22,6 +23,14 @@ service DelegatedIdentity {
     // Subscribe to get local and all federated bundles.
     // The lifetime of the subscription aligns to the lifetime of the stream.
     rpc SubscribeToX509Bundles(SubscribeToX509BundlesRequest) returns (stream SubscribeToX509BundlesResponse);
+
+    // Fetch JWT-SVIDs for workloads that match the given selectors, and
+    // for the requested audience.
+    rpc FetchJWTSVIDs(FetchJWTSVIDsRequest) returns (FetchJWTSVIDsResponse);
+
+    // Subscribe to get local and all federated JWKS bundle.
```
amartinezfayo, line 31:
```suggestion
    // Subscribe to get local and all federated JWKS bundles.
```
```suggestion
    // Subscribe to get local and all federated JWKS bundle.
```

amartinezfayo (commented):
Thank you @loveyana for this contribution! It looks great.
I have only one super minor suggestion in a comment.

amartinezfayo (approved):

evan2645:
I think things will get cumbersome when the workload-to-audience complexity explodes, but I also don't have a better suggestion ... I think in this case, folks are just going to have to manage multiple streams. Maybe it's something we can improve on in the future after some use ... general shape etc looks good to me @amartinezfayo , thank you for the contribution @loveyana!

PR #21, opened 2022-03-09
Introduce an API for key management

evan2645:
Here is my first (err ... maybe third) pass at the key management API that is necessary for https://github.com/spiffe/spire/issues/1934

It is definitely still not done ... and there are some implementation nasties I haven't managed to figure out yet. I'm also not sure if this is generally the right shape or not ... so below are some high level points/problems that I could use some feedback on:
* The bundle API allows for management of all bundles. This local authority API allows for management of local authorities, which have side effects in the bundle (e.g. prepare step or revoke). It is weird to have both with this under-the-cover interaction between them. For now, I have chosen to keep all these operations on this (local authority) api rather than split the operations across a key/authority api and the bundle api. The (major) consequence of doing the latter is that the bundle api would then have to grow identifiers so that individual bundle elements can be addressed. It's also unclear how much of these changes make sense in the context of federated bundles.
* These operations may not do what people think they do under certain circumstances. For example, calling RevokeX509Authority on a server that has an UpstreamAuthority loaded ... would be weird and would not really revoke the authority. So I think we need to somehow limit some or all of these operations to only "root" level servers. This is further complicated by the fact that some "root" level servers may still have an UpstreamAuthority that supports only X.509 (e.g. aws_pca)...
* I think we are missing some safety stories. What happens if you activate a new key then start observing failures in your systems? Similar question around revoke .. we probably want to make these reversible somehow? Lots of implementation questions around that too ...
* JWT doesn't have chaining, which complicates the "root" servers only story. If an intermediate cluster is compromised, and you need to rotate the JWT keys .. either you can run the whole response from this intermediate cluster (i.e. rotate and revoke), or you can split it up (rotate on intermediate cluster, revoke on root cluster). In the former case, we'll need new APIs for downstream servers to send this message up to the root .. and in the latter, it might be hard to correlate which keys etc need touching on intermediates + root sever clusters. In both cases, the divergence in response between JWT and X509 is very confusing and you're probably doing both at the same time. Ideally, the procedures are identical.


proto/spire/api/server/local_authority/v1/local_authority.proto
```diff
@@ -0,0 +1,174 @@
...
+        rpc RevokeX509Authority(RevokeX509AuthorityRequest) returns (RevokeX509AuthorityResponse);
+}
+
+message GetJWTAuthorityStateRequest {}
+
+message GetJWTAuthorityStateResponse {
+        repeated AuthorityState states = 1;
+}
+
+message PrepareJWTAuthorityRequest {}
```
evan2645, line 86: Since we have only two slots per authority type, we can infer which key we want to take the action on based on the type of action. For example, we only ever want to revoke non-active key, and we only ever want to activate a prepared key. In this way, I've managed to sidestep the need to (re)invent key identifiers, however I have still included an opaque slot name for ease of use.

.gitignore
```diff
@@ -1 +1,2 @@
 .build
+*.swp
```
azdagron, line 2: I always neglect adding this because I mercilessly disable the swap file in vim :)

proto/spire/api/server/local_authority/v1/local_authority.proto
azdagron: That makes sense to me.

proto/spire/api/server/local_authority/v1/local_authority.proto
```diff
@@ -0,0 +1,174 @@
...
+
+message RevokeX509AuthorityRequest {}
+
+message RevokeX509AuthorityResponse {
+        AuthorityState revoked_authority = 1;
+}
+
+message AuthorityState {
+        // An opaque name for the "slot" that this authority is housed in
+        string slot = 1;
```
azdagron, line 142: I'm not sure we'll be able to fill this for anything but empty, prepared, and active, particularly if we remove the journal as we've been discussing. My gut feeling says that the slot is an internal detail and probably shouldn't be exposed.

proto/spire/api/server/local_authority/v1/local_authority.proto
```diff
@@ -0,0 +1,174 @@
...
+        // will propagate the new public key cluster-wide.
+        rpc PrepareJWTAuthority(PrepareJWTAuthorityRequest) returns (PrepareJWTAuthorityResponse);
+
+        // ActivateJWTAuthority activates a prepared JWT authority for use,
+        // which will cause it to be used for all JWT signing operations
+        // serviced by this server going forward. If a new JWT authority has
+        // not already been prepared, a FailedPrecondition error will be returned.
+        rpc ActivateJWTAuthority(ActivateJWTAuthorityRequest) returns (ActivateJWTAuthorityResponse);
+
+        // TaintJWTAuthority marks the previously active JWT authority as
```
azdagron, line 23: The "previously active" nomenclature is a little confusing to me. Does that mean you can only perform this action on OLD/EXPIRED authorities? Or does this operation only target the current "active" authority?

proto/spire/api/server/local_authority/v1/local_authority.proto
```diff
@@ -0,0 +1,174 @@
...
+        // being tainted. SPIRE Agents observing an authority to be tainted
+        // will perform proactive rotations of any key material related to
+        // the tainted authority. The result of this action will be observed
+        // cluster-wide.
+        //
+        // If a previously active JWT authority does not exist (e.g. if one
```
azdagron, line 29: I'm not sure we can be in the state, at least with the way the manager is organized. It will automatically activate a prepared authority if there is no active authority....

azdagron (commented):
Thank you very much @evan2645 for putting this together. I left some initial comments. I'm still working the shape through in my mind. I think you are right though to keep bundle stuff separate. I feel that this API is conceptually distinct since not all authorities are represented directly in the bundle.

rturner3:
> I think we are missing some safety stories. What happens if you activate a new key then start observing failures in your systems? Similar question around revoke .. we probably want to make these reversible somehow? Lots of implementation questions around that too ...

Agreed. I think a helpful approach to clarifying these ambiguities would be to come up with a key state diagram that shows how you can get from one state to another, with the arrows in the diagram labeled with the operation(s) required to reach that destination state.

rturner3:
>These operations may not do what people think they do under certain circumstances. For example, calling RevokeX509Authority on a server that has an UpstreamAuthority loaded ... would be weird and would not really revoke the authority. So I think we need to somehow limit some or all of these operations to only "root" level servers. This is further complicated by the fact that some "root" level servers may still have an UpstreamAuthority that supports only X.509 (e.g. aws_pca)...

A couple ideas on this:

- We could potentially return the `UNIMPLEMENTED` status code for methods that are not supported based on Server configuration that cannot be resolved by changing system state via APIs (e.g. configured to use an UpstreamAuthority plugin)
- Perhaps including the word `Root` in the naming of the service or methods could somewhat clarify this ambiguity?

azdagron:
> We could potentially return the UNIMPLEMENTED status code

That's an idea, I need to think some more. Though I'd probably recommend FailedPrecondition v.s. Unimplemented.

proto/spire/api/server/local_authority/v1/local_authority.proto
MarcosDY: I'm the same way, right now we has A or B values, that we use as slotID, based in status, we can avoid this slotID and relly o state, so we can have only one PREPARED | ACTIVE  | OLD, 
I'm not sure about the use of EXPIRED, what is the difference? may we just remove expired authorities?

proto/spire/api/server/local_authority/v1/local_authority.proto
MarcosDY, line 23: we does not have a TAINTED state, how we'll hold that tainted authority? we'll need to keep it from some time until all agents move to the new bundles.
If I'm not missing something the idea of this is to force agents to move and sign new SVIDs

proto/spire/api/server/local_authority/v1/local_authority.proto
```diff
@@ -0,0 +1,174 @@
...
+        rpc TaintX509Authority(TaintX509AuthorityRequest) returns (TaintX509AuthorityResponse);
+
+        // RevokeX509Authority revokes the previously active X.509 authority by
+        // removing it from the bundle and propagating this update throughout
+        // the cluster.
+        //
+        // If a previously active X.509 authority does not exist (e.g. if one
+        // has been prepared but not activated yet), a FailedPrecondition
+        // error will be returned.
+        rpc RevokeX509Authority(RevokeX509AuthorityRequest) returns (RevokeX509AuthorityResponse);
```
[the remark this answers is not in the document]
MarcosDY: an option can be, not removing  from the list of bundles but keep in a `removed` state, and add a new function "ReactivateX509Authority" (same for JWT) that allow activate a bundle that is in a removed || old states this will move actual activated bundle into old state and Activate the chosen one.

To do that we'll need an identifier... maybe the bundle serial number?

Another option is to just use Bundle.AppenBundle with an old bundle... that we'll need to keep in some way so Admins can get it, but that may not work for jwt right?

proto/spire/api/server/local_authority/v1/local_authority.proto
```diff
@@ -0,0 +1,174 @@
+syntax = "proto3";
+package spire.api.server.local_authority.v1;
+option go_package = "github.com/spiffe/spire-api-sdk/proto/spire/api/server/local_authority/v1;local_authorityv1";
+
+// The LocalAuthority service provides a way to manage the signing keys (and
+// related material) of the SPIRE Server exposing it.
+service LocalAuthority {
+        // GetJWTAuthorityState returns the state of all locally configured
+        // JWT authorities.
+        rpc GetJWTAuthorityState(GetJWTAuthorityStateRequest) returns (GetJWTAuthorityStateResponse);
```
MarcosDY, line 10: what is included in get authority state? I have the feeling we must not expose expired (since we are going to cleanup all expired authorities)
I'm not sure when "Prepared" authority can be useful, maybe to know if it is was exposed and want to prepare a new one?

proto/spire/api/server/local_authority/v1/local_authority.proto
```diff
@@ -0,0 +1,174 @@
...
+
+        // PrepareJWTAuthority prepares a new JWT authority for use by
+        // generating a new key and injecting it into the bundle. This action
+        // will propagate the new public key cluster-wide.
+        rpc PrepareJWTAuthority(PrepareJWTAuthorityRequest) returns (PrepareJWTAuthorityResponse);
```
MarcosDY, line 15: what happens with previously prepared bundle? it was never used, so we must just remove it?
and when do you think Prepare a new authority can be useful? in case it was exposed before used?

proto/spire/api/server/local_authority/v1/local_authority.proto
```diff
@@ -0,0 +1,174 @@
...
+
+        // ActivateJWTAuthority activates a prepared JWT authority for use,
+        // which will cause it to be used for all JWT signing operations
+        // serviced by this server going forward. If a new JWT authority has
+        // not already been prepared, a FailedPrecondition error will be returned.
+        rpc ActivateJWTAuthority(ActivateJWTAuthorityRequest) returns (ActivateJWTAuthorityResponse);
```
MarcosDY, line 21: if I follow you here we have 3 functions that allow rotation:
- Activate: it is like regular (and automatic) rotation
- Taint: rotate actual active authority and force mint of new SVIDs on agents, but bundle is not removed
- Revoke: rotate, and DELETE bundle, and force mint new SVIDs.

I think taint/revoke are good for different scenarios where we want to force a rotation, but when `Activate` is used? 
for testing? it will not force SVID rotations, why not to wait for regular rotation?

proto/spire/api/server/local_authority/v1/local_authority.proto
```diff
@@ -0,0 +1,174 @@
...
+                // signing operations.
+                ACTIVE = 3;
+
+                // This is an old authority that was previously in use but is
+                // no longer.
+                OLD = 4;
+
+                // This is an old authority that was previously in use but is
+                // now expired.
+                EXPIRED = 5;
```
MarcosDY, line 169: not sure we  want to expose expired authorities, since we'll automatically remove them

evan2645:
Closing in favor of https://github.com/spiffe/spire-api-sdk/pull/37 thank you @MarcosDY!

PR #22, opened 2022-04-29
Add Mac M1 support

faisal-memon:
Signed-off-by: Faisal Memon <fymemon@yahoo.com>

Adds Mac M1 support to this repo. PR is opened against `next` branch as that seems to have the latest code. Can switch to `main` if thats better. The changes in this PR are:

- Add `arm64` target. M1 identifies as `arm64` instead of `aarch64`
- Use `arch2` to identify `protoc_url`. `arch2` will match for M1 and other arm platforms.
- Update to `protoc` 3.20.1. Older versions don't support M1.
- Regenerated `*.pb.go` files.


azdagron (approved):
Yes, `next` is the right branch for PRs to target and the branch that SPIRE depends on. We pull officially supported things into main and tag from there as part of releasing.

This looks good. Thanks for this :)

PR #25, opened 2022-06-01
Add reattestable bool and AGENT_MUST_REATTEST perm denied

faisal-memon:
Part of https://github.com/spiffe/spire/pull/3031. Contains:

- A `reattestable` bool that gets sent to agent as part of the attestation result. The agent will use this bool to determine if it should reattest or rotate when its svid expires.
- `PermissionDeniedDetails_AGENT_MUST_REATTEST` that gets sent to an agent that attempts to rotate its svid when it should be reattesting.


proto/spire/api/server/agent/v1/agent.proto
```diff
@@ -145,6 +145,9 @@ message AttestAgentResponse {
     message Result {
         // The agent X509-SVID.
         spire.api.types.X509SVID svid = 1;
+
+	// Should the attested agent reattest when its certificate expires
+	bool reattest = 2;
```
azdagron, line 150:
```suggestion
	// Whether or not the attested agent can reattest to renew its X509-SVID
	bool reattestable = 2;
```

proto/spire/api/types/status.proto
```diff
@@ -22,6 +22,8 @@ message PermissionDeniedDetails {
         AGENT_NOT_ACTIVE = 3;
         // Agent has been banned.
         AGENT_BANNED = 4;
+	// Agent attempted to renew SVID, but should reattest instead
+	AGENT_CAN_REATTEST=5;
```
azdagron, line 26:
```suggestion
	AGENT_MUST_REATTEST=5;
```

azdagron (approved):
Thanks, @faisal-memon .

PR #26, opened 2022-07-13
Update dependencies and re-gen protos

rturner3:
Keep dependencies up-to-date to minimize risk of gRPC/protobuf compatibility drift between clients and SPIRE.


go.mod
```diff
@@ -1,9 +1,16 @@
 module github.com/spiffe/spire-api-sdk
 
-go 1.12
+go 1.18
```
azdagron, line 3: Hmm, I think we need to be careful about bumping this. Like go-spiffe, these SDKs are intended to be consumed by organizations that may or may not be on the latest release of Go, and some which are likely pinned to older versions for various reasons....

rturner3: Yeah that's a good point. I narrowed this PR to just include the dep updates and leave out the Go version bump.

azdagron:
Thanks, @rturner3 !

PR #27, opened 2022-07-26
Deprecate ambiguous svids_count field in Agent GetDebug API

rturner3:
Introduce new SVID count fields for each SVID cache in SPIRE Agent. These new fields clarify what exactly is represented in the counts and provide integration tests a way to assert on the state of the various Agent in-memory caches.

Relates to spiffe/spire#3256


MarcosDY:
this looks fine to me, I have a question, now that we are thinking on introducing a `cache size` maybe it works to add a subscribers_count to debug endpoint, what do you think about that?

rturner3:
> this looks fine to me, I have a question, now that we are thinking on introducing a `cache size` maybe it works to add a subscribers_count to debug endpoint, what do you think about that?

By subscribers, do you mean cache subscribers or Workload API streams (FetchX509SVIDs)? Either way, IMO that seems out of scope of these changes.

MarcosDY (approved):
LGTM!!!

PR #29, opened 2022-09-21
Adds support for X509 and JWT specific SVID TTLs

dennisgove:
This change is in reference to a new SPIRE feature discussed in https://github.com/spiffe/spire/issues/2700

Signed-off-by: Dennis Gove <dgove1@bloomberg.net>


azdagron:
This looks great @dennisgove. I need to update the CONTRIBUTING docs on this, but can you target this PR against the `next` branch? We only pull things into main when we are cutting a release.

Also, would you mind documenting the `ttl` field with a notice of deprecation? something along the lines of, `This field is deprecated in favor of the x509_svid_ttl and jwt_svid_ttl fields and will be removed in a future release`.

dennisgove:
Thanks @azdagron. I've rebased off `next` and moved the PR to that branch. Have also added a deprecation notice on the `ttl` field. I'll add a similar deprecation notice in the `spire` codebase.

azdagron:
Awesome. Looks like `make` needs to be run again to regen the code after adding that comment.

azdagron (approved):
Thanks, @dennisgove !

PR #35, opened 2023-01-24
Add hint filter field in Filter message

guilhermocc:
Add hint field in `ListEntriesRequest` `Filter` message. This field can be used for filtering registration entries by Hint in a list entries request.


proto/spire/api/server/entry/v1/entry.proto
```diff
@@ -60,6 +60,7 @@ message ListEntriesRequest {
         spire.api.types.SPIFFEID by_parent_id = 2;
         spire.api.types.SelectorMatch by_selectors = 3;
         spire.api.types.FederatesWithMatch by_federates_with = 4;
+        string by_hint = 5;
```
azdagron, line 63: Hmm, I wonder if this should be a google.protobuf.StringValue? Is there value in filtering by entries that have no hint set?

guilhermocc: Hey @azdagron, I'm not sure about your question, but I got your point here; what do you mean by the value that has no hint set?  It makes sense to only apply the filter in a hint if a value `not empty` is provided. I didn't know `google.protobuf.StringValue`, but you suggested it for being able to pass a null value here when we don't want the filter to be applied, right? in this case, why not use the `optional` label?

azdagron: All fields are optional in proto3 and the generated code makes it impossible to distinguish between "unset" and "set with the zero value" for all primitive types. StringValue puts a string in a message so you can distiguish.

guilhermocc: Awesome, thank you for the clarification 😄

guilhermocc: [1d55464](https://github.com/spiffe/spire-api-sdk/pull/35/commits/1d55464706951be29c8aea44d45127b382e436fd)

azdagron (approved):
\o/

PR #37, opened 2023-02-20
Introduce local authority api

MarcosDY:
Update Evan's proposal with a the introduction of some required changes found on POC implementation, 
and add some simplifications.

Fixes: https://github.com/spiffe/spire/issues/3885


proto/spire/api/server/agent/v1/agent.proto
```diff
@@ -190,3 +197,13 @@ message AgentX509SVIDParams {
     // ignored. The agent X509-SVID attributes are determined by the server.
     bytes csr = 1;
 }
+
+message PushStatusRequest {
+    // Required. Serial number of the bundle used to sign Agent identity.
+    string authority_serial = 1;
```
MarcosDY, line 203: The idea of this serial is to keep on server database the current serial number used for registered agent.
In POC I didn't implement this, so we'll require some changes to access this data (maybe on agent endpoint) 
We can avoid this on initial implementation too...

proto/spire/api/server/agent/v1/agent.proto
```diff
@@ -190,3 +197,13 @@ message AgentX509SVIDParams {
...
+}
+
+message PushStatusResponse {
+    // Keys that are no longer secure and must be rotated.
+    repeated bytes taintedKeys = 1;
```
MarcosDY, line 208: The intention here is to return the list of tainted keys (JWT or X.509) so agent can react renewing all affected SVIDs (agent SVID, cached workload X.509 SVIDs, cached JWT SVIDs)

proto/spire/api/server/localauthority/v1/localauthority.proto
```diff
@@ -0,0 +1,189 @@
...
+
+message ActivateJWTAuthorityRequest {}
+
+message ActivateJWTAuthorityResponse {
+        AuthorityState activated_authority = 1;
+}
+
+message TaintJWTAuthorityRequest {
+        // The key ID of the local authority to taint
+        string key_id = 1;
```
MarcosDY, line 103: the idea is to use "old"  authority if no key_id is provided, and X.509 implementation requires a public key... 
I'm worried about the difference about both, we may use public key here too..
but our endpoints returns a key_id so I think this can be more natural to use that key id (and our JWT SVIDs has the key ID too)

proto/spire/api/server/localauthority/v1/localauthority.proto
```diff
@@ -0,0 +1,189 @@
...
+
+message ActivateX509AuthorityRequest {}
+
+message ActivateX509AuthorityResponse {
+        AuthorityState activated_authority = 1;
+}
+
+message TaintX509AuthorityRequest {
+        // The public key of the local authority to taint
+        bytes public_key  = 1;
```
MarcosDY, line 139: an alternative to public key is to use Serial number,
and this proposal has the PushStatus that contains the serial number of the bundle used to sign Agent SVID too, 
I was worried about 2 bundles using the same Serial number, but I have the feeling I overkill this...

proto/spire/api/server/svid/v1/svid.proto
```diff
@@ -125,6 +126,9 @@ message NewDownstreamX509CAResponse {
 
     // X.509 authorities (DER encoded).
     repeated bytes x509_authorities = 2;
+
+    // X.509 authorities with metadata
+    repeated spire.api.types.X509Certificate x509_authorities_with_metadata = 3;
```
MarcosDY, line 131: this is too be sure that on initial downstream call, we get the list of tainted keys

proto/spire/api/server/localauthority/v1/localauthority.proto
```diff
@@ -0,0 +1,179 @@
...
+message RevokeX509AuthorityRequest {
+        // The public key of the local authority to revoke
+        bytes public_key  = 1;
+}
+
+message RevokeX509AuthorityResponse {
+        AuthorityState revoked_authority = 1;
+}
+
+message AuthorityState {
```
evan2645, line 155: I think we need key ID as part of this message. An operator will first want to list the authorities, and then will want to taint one, and to taint you need the key ID.

proto/spire/api/server/agent/v1/agent.proto
MarcosDY: agree, maybe we can create an issue on spire to track all the work about persisting and recovering this information with some APIs

proto/spire/api/server/agent/v1/agent.proto
MarcosDY: for now it is a simple list, and we verify if that key was used for jwt or x509, and taint any of them, 
Maybe I overthink this, and we can just create 2 lists and didn't verify if the key was used by only one of our bundles

proto/spire/api/server/localauthority/v1/localauthority.proto
MarcosDY: ok, so move to taint using public keys on both looks like the correct call.

proto/spire/api/server/localauthority/v1/localauthority.proto
MarcosDY: yeah that was what worry me...
a simple key ID is easier to use, than a public key...
but our GET are getting that public keys, and at the same time,  it is easy to get the public key from the affected bundle if you have it.. 
So maybe it is safer to keep public key for now... and as you said nothing prevents to get the same key in multiple bundles (that is why I added a single TaintedKey list to agent, so we can be sure we taint any affected bundle, no matter if it was a JWT or X.509)

proto/spire/api/server/svid/v1/svid.proto
MarcosDY: all of them has different uses, 
`NewDownstreamX509CAResponse`: upstream authority is the one communicating with the Upstream Authority, so that plugin can call getBundle from Upstream and get the list of affected bundles, and propagate that information to downstream.
so we can taint the Downstream bundle that was signed by an unsecure authority.

So this is to get the bundle when communication between plugin and agent started after that we can rely on getbundle to verify if there are updates

MarcosDY: in the end the list of bundles and state is returned to ca manager (on server)  for processing and rotate if required

proto/spire/api/server/agent/v1/agent.proto
```diff
@@ -58,6 +58,13 @@ service Agent {
     //
     // The caller must be local or present an admin X509-SVID.
     rpc CreateJoinToken(CreateJoinTokenRequest) returns (spire.api.types.JoinToken);
+
+    // Pushes Agent status, relevant information is returned to agent to keep it 
+    // updated on server changes.
+    //
+    // The caller must present an active agent X509-SVID, i.e. the X509-SVID
+    // returned by the AttestAgent or the most recent RenewAgent call.
+    rpc PushStatus(PushStatusRequest) returns (PushStatusResponse);
```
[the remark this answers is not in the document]
MarcosDY: agree, that sounds like an stream, I'm rename to PostStatus

proto/spire/api/server/localauthority/v1/localauthority.proto
MarcosDY: yeah, I was thinking on adding that based is what we decide about taint and revoke, if unify using public key on both or use key_id for one and public key for another one,
based in what we choose here is what we must return

proto/spire/api/server/localauthority/v1/localauthority.proto
MarcosDY: Using SKID looks like something possible, of course it will be only for endpoints but internally I think we must keep tainting Public keys, and another endpoints may propagates pubic keys

proto/spire/api/server/agent/v1/agent.proto
```diff
@@ -193,3 +200,11 @@ message AgentX509SVIDParams {
     // ignored. The agent X509-SVID attributes are determined by the server.
     bytes csr = 1;
 }
+
+message PostStatusRequest {
+    // Required. Serial number of the bundle currently being served by the agent
+    int64 current_bundle_serial = 1;
```
MarcosDY, line 206: may we use uint64 here?

proto/spire/api/server/svid/v1/svid.proto
evan2645: I still wonder if there's some way around having to do this .. I spent a little time looking at the upstream authority plugin code, and it appears that we get root keys from JWT response, X509 response, and then combine them to form the bundle that should be propagated downstream. And then that causes us to have to poke these taint bits through

One option could be to change the plugin logic to load from the upstream bundle API sooner, and use it purely as the source of truth (rather than use JWT + X509 roots as source of truth on boot, then change over to bundle API). That way, the only place we communicate taint data is through the bundle.

I don't think this is blocking though, just an observation.

evan2645 (approved):
thanks for your patience @MarcosDY ❤️

proto/spire/api/server/localauthority/v1/localauthority.proto
```diff
@@ -0,0 +1,185 @@
...
+        repeated AuthorityState states = 1;
+}
+
+message PrepareJWTAuthorityRequest {}
+
+message PrepareJWTAuthorityResponse {
+        AuthorityState prepared_authority = 1;
+}
+
+message ActivateJWTAuthorityRequest {}
```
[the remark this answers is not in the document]
MarcosDY: I was thinking the same thing, but I was not sure about implementing that on the first version, 
I agree it can be useful, and we can implement this, with some changes on code (nothing too bad) 
So I'll add it now, and we can discuss on implementation.

evan2645 (approved):
🪨 🥧 

thank you Marcos ❤️ looks great

PR #38, opened 2023-03-13
Expose created_at field in Entry service

guilhermocc:
Following the specification defined in https://github.com/spiffe/spire/issues/745, and the need for using this field in the tie-breaking logic that will be develop for ensuring hint uniqueness in https://github.com/spiffe/spire/issues/2588, this PR exposes the created_at field in Entry service.


rturner3 (approved):
LGTM!

PR #43, opened 2023-06-20
Get local authorities to attributes

MarcosDY:
Update get local authorities to return authorities by state instead of a list


proto/spire/api/server/localauthority/v1/localauthority.proto
```diff
@@ -128,7 +140,19 @@ message RevokeJWTAuthorityResponse {
 message GetX509AuthorityStateRequest {}
 
 message GetX509AuthorityStateResponse {
-        repeated AuthorityState states = 1;
+        // Autority currently beign used for signing operations.
+        AuthorityState Active = 1;
```
azdagron, line 144:
```suggestion
        AuthorityState active = 1;
```

proto/spire/api/server/localauthority/v1/localauthority.proto
```diff
@@ -83,7 +83,19 @@ service LocalAuthority {
 message GetJWTAuthorityStateRequest {}
 
 message GetJWTAuthorityStateResponse {
-        repeated AuthorityState states = 1;
+        // Autority currently beign used for signing operations.
```
azdagron, line 86:
```suggestion
        // Autority currently being used for signing operations.
```

proto/spire/api/server/localauthority/v1/localauthority.proto
```diff
@@ -128,7 +140,19 @@ message RevokeJWTAuthorityResponse {
 message GetX509AuthorityStateRequest {}
 
 message GetX509AuthorityStateResponse {
-        repeated AuthorityState states = 1;
+        // Autority currently beign used for signing operations.
```
azdagron, line 143:
```suggestion
        // Authority currently beign used for signing operations.
```

proto/spire/api/server/localauthority/v1/localauthority.proto
```diff
@@ -83,7 +83,19 @@ service LocalAuthority {
...
+        // Authority added on bundle but is not used yet.
+        AuthorityState prepared = 2;
+
+        // Authority in that was previously used for signing operations,
+	// but it is not longer.
+        AuthorityState old = 3;
+
+        // Authority without state, This in an error condition and should 
+	// never be encountered.
+        AuthorityState unknown = 4;
```
azdagron, line 98: I don't see a good reason to return this. The caller can't do anything with it. If this situation was ever encountered, server logs would probably suffice?

proto/spire/api/server/localauthority/v1/localauthority.proto
```diff
@@ -128,7 +140,19 @@ message RevokeJWTAuthorityResponse {
...
```
azdagron, line 155: Same comment as the unknown state in GetJWTAuthorityStateResponse

azdagron (commented):
I don't see a reason to keep the "State" field in the AuthorityState message with these fields present. I also wonder if tainted should have its own field in the Get(X509|JWT)AuthorityState response (repeated AuthorityState tainted).

What's the criteria for something being old? How does that differ from prepared? And how do you distinguish? Is old just all non-active, non-tainted keys that are older than the active key?

Let's say we have the following keys, ordered by descending creation date.

K5
K4
K3
K2
K1

Anything newer than the active key set, that is also not tainted, is "prepared". So If K3 is "active", then K4 and K5 is "prepared". Anything older than the active key set, that is also not tainted, is "old" (well maybe... it might never have been used), so K2 and K1 are "old".

K5 <-- prepared
K4 <-- prepared
K3 <-- active
K2 <-- old
K1 <-- old

Now, we might constrain these buckets and say that there can be at most one "prepared" key, and only one "old" key by the way we construct our RPCs and state changes.

Seems like there can be any number of tainted keys.

MarcosDY:
> What's the criteria for something being old? How does that differ from prepared? And how do you distinguish? Is old just all non-active, non-tainted keys that are older than the active key?

Old was previously activate, or just a Prepared key that moved into Old state in case we force to have a new Prepared key, 
we can no longer rely on dates, since a use can play with running PrepareKey multiple times, and not forcing the rotation of the activate key.
So in SPIRE we must be sure we only have:
1 active
1 prepared
n old (on journal), but we must return only the newest OLD
n tainted keys (on datastore) but in the end we can only return 1 that is in OLD slot... this part is tricky...

proto/spire/api/server/localauthority/v1/localauthority.proto
MarcosDY: agree, there is no reason to return unknown authorities, even we must never get into this situation

azdagron (approved):

evan2645 (approved):
Thanks @MarcosDY and @azdagron ❤️

PR #47, opened 2023-09-04
Force rotation proposal for upstream authority

MarcosDY:
When using Upstream authorities, intermediate certificates are not part of bundle, because bundle is the upstream authority CA, 
because of that we need to provide a way to propagate the list of tainted keys in case of X.509 authorities, (in JWT it is not required becase our JWT authority is always in datastore)


proto/spire/api/types/bundle.proto
```diff
@@ -18,14 +18,20 @@ message Bundle {
 
     // The sequence number of the bundle.
     uint64 sequence_number = 5;
+
+    // The list of X.509 authorities that are no longer secure to use, 
+    // and must be rotated.
+    repeated X509TaintedKey x509_tainted_keys = 6;
```
MarcosDY, line 24: It is only required for x509 authorities, but it feels strange to not have the same list for JWT authorities, what do you think?

PR #49, opened 2023-11-05
New SyncAuthorizedEntries RPC

azdagron:


proto/spire/api/server/entry/v1/entry.proto
```diff
@@ -46,6 +46,21 @@ service Entry {
     // The caller must present an active agent X509-SVID. See the Agent
     // AttestAgent/RenewAgent RPCs.
     rpc GetAuthorizedEntries(GetAuthorizedEntriesRequest) returns (GetAuthorizedEntriesResponse);
+
+
```
MarcosDY, line 50:
```suggestion

```

proto/spire/api/server/entry/v1/entry.proto
```diff
@@ -46,6 +46,21 @@ service Entry {
...
+    // Syncs authorized entries down the caller. The caller controls which
```
MarcosDY, line 51: Syncs authorized entries down 'to' the caller.?

proto/spire/api/server/entry/v1/entry.proto
```diff
@@ -46,6 +46,21 @@ service Entry {
...
+    // entries the server sends down full details for. The flow is as follows:
+    // 1. Caller opens up sync stream
+    // 2. Server determines authorized entries for caller:
+    //    - If there are less entries than a server-determined page size, go to (5).
```
MarcosDY, line 55: it is going to 5? or 4?
this means that we get full if we return entries < pagesize?

proto/spire/api/server/entry/v1/entry.proto
```diff
@@ -176,3 +191,24 @@ message GetAuthorizedEntriesResponse {
...
+
+    // IDs of the entries to fetch in full. Sent by the client in response to
+    // a sparse entry.
+    repeated string ids = 2;
+}
+
+message SyncAuthorizedEntriesResponse {
+    // Full or sparse entries, depending on the sparse field. See
+    // SyncAuthorizedEntries for details.
+    repeated spire.api.types.Entry entries = 1;
```
MarcosDY, line 207: I'm ok with this, but, may we use oneof? and return another type of "entries"? instead of depending of an sparse?

proto/spire/api/server/entry/v1/entry.proto
azdagron: Thanks :)

proto/spire/api/server/entry/v1/entry.proto
azdagron: To (5). If there aren't many entries, it's not worth trying to do the "diff". The server just sends the full non-sparse set to the agent.
The outcome is that there is no noticeable difference from a network standpoint for existing deployments whose authorized entry sets are smaller than the page size.

MarcosDY (approved):
Looks good, 
But I have a question, what happens when `SyncAuthorizedEntriesRequest` gets an ID that the agent is not authorized to return? is it ignored?

azdagron:
Ids not in the authorized set are ignored.

PR #52, opened 2024-01-12
Support filters in entry & agent command

FedeNQ:
# Description
* Make required changes in entry.proto to support filtering while using count/show command.
* Make required changes in agent.proto to support filtering while using count/list command.


amartinezfayo (approved):
LGTM!

Issue #41, opened 2023-05-09
Support a prune RPC in the SPIRE Server's Agent API


evan2645:
SPIRE PR https://github.com/spiffe/spire/pull/3982 added a spire-server CLI command for pruning old agents out of the database. That work does client-side filtering of agents to determine which need pruning. During review, this was flagged and we agreed that the prune logic is best handled server-side.

Introduce and implement a new RPC to support this server-side prune feature

nweisenauer-sap:
Being able to filter `ListAgents` by `X509SvidExpiresAt` would already be very helpful in improving performance.

Additionally, @evan2645, if the whole [prune/purge command #3982](https://github.com/spiffe/spire/pull/3982) will get its own RPC, what do you think about giving it a flag/parameter to force deletion of expired agents regardless of their `CanReattest` property? Maybe even a parameter to specify the limit / the maximum number of entries to be pruned in one run, just to prevent unexpected load. 
Please consider that there is currently no other way of deleting stale entries from TOFU-based plugins (these may be expired for months or years and eventually be cleaned up with manual, risky SQL queries on the DB).

PR #53, opened 2024-01-15
Allow specifying "ByExpiresBefore" in the "ListAgentsRequest_Filter"

nweisenauer-sap:
Allowing to filter agents "ByExpiresBefore" on the server side allows for a more efficient implementation of pruning expired agent entries.
This is a prerequisite for a PR in spiffe/spire to allow "ListAgents" to filter "ByExpiresBefore".

**Which problem does this solve for us?**
This allows users to implement their own cleanup logic in a more efficient way by fetching only expired agents.

Ref:
- #41
- [Feature Request: Filter ListAgent using the value of X509SvidExpiresAt #2367](https://github.com/spiffe/spire/issues/2367)


nweisenauer-sap:
Seems that #52 has implemented the same thing and has been merged, so this PR can be closed?

PR #81, opened 2025-10-27
Add new WIT-related RPCs

arndt-s:
Related to #78 #79.

This PR mainly adds new RPCs for the WIT-SVID. 

Design decisions
* WIT Authorities is identified using key id like JWKS. However, I'm not sure if this cases conflicts with the JWT authority identifier internally in SPIRE?
* I decided to format the public key required when requesting a WIT-SVID as an byte array with ASN1 DER encoding. The CSR as in the X509-SVID doesn't make sense for the WIT, while for JWT no public key is required. Alternatively, to keep the technology aligned a JWKS carrying the private key is also possible. Let me know in the comments.
* Instead of a `NewWITSVID` rpc in the SVID API, I directly went with `BatchNewWITSVID` like the X.509 SVID does. Since there's no audience this makes more sense to me.


arndt-s:
Oh, I just noticed that this is a duplicate of #80 😔

sorindumitru:
I think we can close these now since we have WIT RPCs on main.

PR #89, opened 2026-01-26
fix: Updates go_url to correct location

jsnctl:
Running a fresh `make` was failing for me as follows:

```
💤 spire-api-sdk/ make                                                                            (main)
go_url: https://storage.googleapis.com/golang/go1.23.0.linux-amd64.tar.gz
Installing go1.23.0...
curl: (22) The requested URL returned error: 403

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
make: *** [Makefile:106: go-check] Error 2
```

[It appears as if the canonical location for these dependencies has recently been updated](https://github.com/actions/setup-go/issues/688), and the `https://storage.googleapis.com/golang/*` repository no longer applies. See also: https://github.com/actions/go-versions/pull/127

This fix updates the `go_url` value to the new `go.dev/dl` URL (as outlined in the above issues)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ go_version_full := $(shell cat .go-version)
 go_version := $(go_version_full:.0=.0)
 go_dir := $(build_dir)/go/$(go_version)
 go_bin_dir := $(go_dir)/bin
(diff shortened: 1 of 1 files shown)


sorindumitru:
Sorry we missed this, but it should be fixed now on main.

PR #90, opened 2026-01-26
feat: Adds `wit_svid_ttl`

jsnctl:
Enabling protobuf change for https://github.com/spiffe/spire/issues/6535
Adding `wit_svid_ttl` field to entry message (equivalent to existing X.509-SVID and JWT-SVID TTLs)

### DIFF
--- a/proto/spire/api/types/entry.proto
+++ b/proto/spire/api/types/entry.proto
@@ -66,7 +66,7 @@ message Entry {
     int64 created_at = 15;
 
     // The time to live for WIT-SVID identities issues for this entry (in seconds), overttide ttl if set.
-    int64 wit_svid_ttl = 16;
+    int32 wit_svid_ttl = 16;
 }
 
 // Field mask for Entry fields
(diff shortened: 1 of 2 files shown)


proto/spire/api/types/entry.proto
```diff
@@ -87,6 +87,9 @@ message Entry {
         bool jwt_svid_include_jti = 2;
     }
     optional AdditionalAttributes additional_attributes = 16;
+
+    // The time to live for WIT-SVID identities issues for this entry (in seconds), overttide ttl if set.
```
amartinezfayo, line 91: I think there are two typos here: `issues` should be `issued`, and `overttide` should be `overrides`. The `overrides ttl` phrasing is also inherited from jwt_svid_ttl's legacy field. Would `The server default will be used if unset`, as in MintWITSVIDRequest, be more accurate?

proto/spire/api/types/entry.proto
```diff
@@ -135,4 +138,7 @@ message EntryMask {
 
     // additional_attributes field mask
     bool additional_attributes = 16;
+
+    // wit_svid_ttl mask
```
amartinezfayo, line 142: nit: every other mask here reads `<field> field mask`. Should we make this `wit_svid_ttl field mask` to match?

amartinezfayo (commented):
Thank you @jsnctl! We should target this PR to the `next` branch, instead of main.

amartinezfayo (approved):

PR #92, opened 2026-02-16
Add Broker API

arndt-s:
I wasn't sure if `next` is the correct branch here. Adds SPIFFE Broker API to the spire-api-sdk.

We could discuss putting it in go-spiffe too, but this API is not "consumer" facing so I don't think that would be the correct place and this here is more suitable.

SPIFFE PR: https://github.com/spiffe/spiffe/pull/340
SPIRE PR: https://github.com/spiffe/spire/pull/6594


matheuscscp:
Superseded by: #96

sorindumitru:
This went in through `go-spiffe`

PR #84, opened 2025-12-26
Add JWT-SVID audience policy fields to Entry type

srikalyan:
## Summary

Adds support for per-audience JWT-SVID policy configuration to the Entry type.

### New JWTSVIDAudiencePolicy enum with three modes:
- **DEFAULT (0)**: No JTI claim, caching enabled. Backwards compatible behavior.
- **AUDITABLE (1)**: JTI claim included for audit trails, caching still enabled.
- **UNIQUE (2)**: JTI claim included, caching disabled. Each request gets a fresh token.

### New Entry fields:
- `jwt_svid_default_audience_policy`: Default policy for audiences not explicitly configured
- `jwt_svid_audience_policies`: Map of audience → policy for per-audience overrides

### New EntryMask fields:
- `jwt_svid_default_audience_policy`
- `jwt_svid_audience_policies`

## Related

This SDK change supports the SPIRE server implementation in https://github.com/spiffe/spire/pull/6514

Fixes https://github.com/spiffe/spire/issues/6043

PR #95, opened 2026-04-21
Add jwt_svid_include_jti to Entry.AdditionalAttributes

srikalyan:
## Summary

Adds a new `bool jwt_svid_include_jti` field to `Entry.AdditionalAttributes` (field number 2, scoped inside the nested sub-message introduced by #88).

When set to true on a registration entry, SPIRE:
- Includes a freshly generated UUID in the `jti` claim of every JWT-SVID minted for the entry.
- Bypasses the agent's JWT-SVID cache so every request yields a new token.

This supports use cases that require per-token uniqueness, such as replay protection or auditing.

## Background

This replaces the prior approach (an enum + per-audience policy map, previously proposed in #84) with a single boolean, following upstream feedback from @sorindumitru. The corresponding SPIRE PR is spiffe/spire#6514.

## Test plan

- [x] `make generate` regenerates `entry.pb.go` cleanly
- [x] Consuming SPIRE branch compiles and its full test suite passes against this SDK


sorindumitru (approved):
LGTM, thanks @srikalyan. I'll also have a look at the SPIRE PR

srikalyan:
Thank you @sorindumitru can we merge this, becomes easier on the implementation PR.

PR #96, opened 2026-04-27
Add Broker API protos

matheuscscp:
Supersedes: #92 

SPIFFE PRs: https://github.com/spiffe/spiffe/pull/340 https://github.com/arndt-s/spiffe/pull/1
SPIRE PR: https://github.com/spiffe/spire/pull/6915


proto/spire/api/server/svid/v1/svid.proto
```diff
@@ -27,6 +28,24 @@ service SVID {
...
+    //
+    // The caller must present an active agent X509-SVID.
+    rpc MintBrokerObjectX509SVID(MintBrokerObjectX509SVIDRequest) returns (MintBrokerObjectX509SVIDResponse);
+
+    // Mints a JWT-SVID for a Kubernetes object referenced through the
+    // SPIFFE Broker API. Same flow as MintBrokerObjectX509SVID. The TTL is
+    // determined by the server's default_jwt_svid_ttl configuration.
+    //
+    // The caller must present an active agent X509-SVID.
+    rpc MintBrokerObjectJWTSVID(MintBrokerObjectJWTSVIDRequest) returns (MintBrokerObjectJWTSVIDResponse);
```
sorindumitru: I think generally the agent would either have the an X509-SVID for the workload or will just just call get one through `BatchNewX509SVID` (it does this asynchronously). I don't think we need these extra APIs here, but if you think we need them, let us know why.

matheuscscp: @sorindumitru Thanks for the review.

Does SPIRE have a regular dev meeting? I have a ton of things to discuss for the implementation of the Broker API, this is not the only unusual thing you're gonna see from what I'm about to propose...

sorindumitru: There's a meeting every Tuesday, see https://github.com/spiffe/spiffe/tree/main/community/sig-spire#spire-contributor-sync

I think today we might have something already on the agenda, if the requester can still make it, but if you have time you can join in case we have enough time for this too. Otherwise, you can let us know on the SPIFFE slack, in #spire, that you want to discuss this and we can put it on the agenda.

matheuscscp: Removed this for now

matheuscscp:
Superseded by https://github.com/spiffe/go-spiffe/pull/388

